PR #7, opened 2019-07-13
problem 1 - please review

hrshah5:


Problem1.java
```diff
@@ -0,0 +1,19 @@
...
+                if(prices[i]<prices[j]){
+                    if (prices[j] - prices[i] > maxProfit){
+                        maxProfit = prices[j] - prices[i];
+                    }    
+                }
+            }
+        }
+        return maxProfit;
+    }
+}
```
ashmeetdeol19: Go for the optimal Solution (Kadanes Algorithm)
And Also submit Problem 2.

PR #64, opened 2020-01-08
Adding solutions to DP-3 - Please review

poojakittur:


Problem_1.java
```diff
@@ -0,0 +1,30 @@
...
+public class Problem_1 {
+
+public int deleteAndEarn(int[] nums) {
+        
+        if(nums.length == 0 || nums == null){
+            return 0;
+        }
+        
+        int values[] = new int[20000];
+        for(int num : nums){
```
super30admin: Take the max value from nums array as basis of length

Sample.java
```diff
@@ -1,5 +1,11 @@
 // Time Complexity :
+//Problem_1 : O(max(nums))
+//Problem_2 : O(m*n)
+
```
super30admin: Please recheck the space for delete and earn here

poojakittur: Since we are using auxiliary space(Array) of max(nums) length, would not the space complexity for delete and earn be O(max(nums)) ?

super30admin: My bad I meant for minimum falling path ...since we are mutating same array

PR #69, opened 2020-02-14
DP - 3 @nikrasiya

Bran99:
@nikrasiya

### DIFF
--- a/delete-and-earn.js
+++ b/delete-and-earn.js
@@ -1,0 +1,45 @@
+// Time Complexity : O(NlgN)
+// Space Complexity : O(N)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : Yes, I couldn't figure it out so I read the discussion/solution to understand it.  Still need the class tomorrow to get the concept.
+
+/**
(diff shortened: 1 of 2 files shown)


delete-and-earn.js
```diff
@@ -0,0 +1,45 @@
+// Time Complexity : O(NlgN)
+// Space Complexity : O(N)
```
super30admin: Please review the time again. We are just iterating over dp array

delete-and-earn.js
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * @param {number[]} nums
+ * @return {number}
+ */
+var deleteAndEarn = function(nums) {
+    // Instantiate an array
+    // of length 10001 (max input size) and fill it with 0
+    const count = new Array(10001).fill(0);
+    let curr = 0,
```
super30admin: Taking an array of 10000 is not recommended . We have committed a space of 10000*4 bits. Rather we can take just the max number among our given inputs

super30admin:
Good Job with problem #2.Please review inline comments

PR #72, opened 2020-02-15
DP-3 completed

Mohan110594:
@nikrasiya

### DIFF
--- a/Delete_Earn.py
+++ b/Delete_Earn.py
@@ -1,0 +1,32 @@
+#// Time Complexity : O(maximum element in the given array nums)
+#// Space Complexity : O(maximum element in the given array nums)
+#// Did this code successfully run on Leetcode : Yes
+#// Any problem you faced while coding this : No
+
+
+#// Your code here along with comments explaining your approach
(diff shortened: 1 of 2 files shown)


Delete_Earn.py
```diff
@@ -0,0 +1,32 @@
+#// Time Complexity : O(maximum element in the given array nums)
+#// Space Complexity : O(maximum element in the given array nums)
```
super30admin: Good. Both are correct

Delete_Earn.py
```diff
@@ -0,0 +1,32 @@
...
+    def deleteAndEarn(self, nums):
+        """
+        :type nums: List[int]
+        :rtype: int
+        """
+        if nums == None or len(nums) == 0:
+            return 0
+        
+        out=[0]*(max(nums)+1)
+
```
super30admin: Don't do this for matrixes. For array `[0]*(max(nums)+1)`is fine. Remind me to cover it in class one more time. We had this discussion in 2nd class.

super30admin:
Good Job . Please review inline comments

PR #74, opened 2020-02-16
after class solutions - please review complexities

pbandyop:
@Lohith-Aswathappa


Lohith-Aswathappa:
Good job Payel!

Please add comments to second problem 2 as well.

PR #78, opened 2020-02-19
Finished DP-3, please review

sowmyadvn:
@BhushanVasisht please review the PR for DP-3 and let me know if anything can be improved. Thanks

### DIFF
--- a/DeleteAndEarn_740.java
+++ b/DeleteAndEarn_740.java
@@ -48,3 +48,31 @@ class Solution {
         return Math.max(skip,choose);
     }
 }
+
+class Solution {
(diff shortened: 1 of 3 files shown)


BhushanVasisht:
Good Job. Please check which code you are submitting before commit. There is an unrelated peoblem in this commit.

BhushanVasisht:
Good Job. Please check which code you are submitting before commit. There is an unrelated peoblem in this commit.

sowmyadvn:
Hi @BhushanVasisht.. Thank you for reviewing. Even if min cost tickets is not in DP-3, it was discussed in class. That’s why submitted that too.

PR #84, opened 2020-03-26
Solution

radhika-github:


super30admin:
Correct. Kindly challenge yourself more. How would you write the brute force of this problem?

super30admin:
Correct. Kindly challenge yourself more. How would you write the brute force of this problem?

PR #87, opened 2020-03-27
Completed Problem 1

arcsoftech:


devashishnyati:
Good job Arihant!
Nice work with the brute force approach! What is the time and space complexity for the brute force solution?
Also, submit the other problem later.

PR #88, opened 2020-03-27
Completed the Min Falling Path problem: Please review

KBangera:


devashishnyati:
Good job Karthik!
Please submit the other problems.

KBangera:
Attempted the delete and earn but it's the wrong algo for a bigger input. Will update the solution after today's class.

PR #94, opened 2020-03-29
Completed DP3 Please Review

kannan-mrugapphan:
Passes all test cases in leetcode
Added the code for minimum cost tickets too

### DIFF
--- a/DeleteAndEarn.java
+++ b/DeleteAndEarn.java
@@ -1,0 +1,36 @@
+//space - O(n) - for support and result arrays
+//time - O(max(n, max(elements in list)))
+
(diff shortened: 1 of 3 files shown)


DeleteAndEarn.java
```diff
@@ -0,0 +1,36 @@
+//space - O(n) - for support and result arrays
+//time - O(max(n, max(elements in list)))
```
devashishnyati: It should be O(n + max(elements in list))

devashishnyati:
Good job Kannan!
Please submit the Brute Force solutions also.

PR #99, opened 2020-05-02
Update Sample.java. Please review DP-3

Abinashi7:


iivaishnavii:
Good job,Abinashi! I notice that you wrote the optimal solution in patch-2. Good to see that you have used both brute force and optimal solution.

Abinashi7:
> Good job,Abinashi! I notice that you wrote the optimal solution in patch-2. Good to see that you have used both brute force and optimal solution.

Thank you!

PR #101, opened 2020-05-02
Completed

namrataKasar:


Sample.java
```diff
@@ -1,7 +1,63 @@
-// Time Complexity :
-// Space Complexity :
-// Did this code successfully run on Leetcode :
-// Any problem you faced while coding this :
 
+class Solution {
 
-// Your code here along with comments explaining your approach
+    // Time Complexity : O(N)
+    // Space Complexity : O(1) constant space of 10001 size array
```
sowmyadvn, line 5: This is correct in this case as we have 10000 as upper bound, but if max value is not mentioned/ is considerably higher, it should be considered auxiliary space

sowmyadvn:
Good job Namrata! Both the solutions and time/space complexities are correct

sowmyadvn:
Good job Namrata! Both the solutions and time/space complexities are correct

PR #111, opened 2020-05-07
[Completed DP-3] Please review

rajasree-r:


deleteEarn.py
```diff
@@ -0,0 +1,21 @@
+# Time Complexity : O(N log N) where N = len(nums)
```
sowmyadvn, line 1: Is this because of collections.Counter() method? Because I have read otherwise: https://stackoverflow.com/questions/42461840/what-is-the-time-complexity-of-collections-counter-in-python

sowmyadvn:
Good job Rajashree! Please review in-line comment

sowmyadvn:
Good job Rajashree! Please review in-line comment

deleteEarn.py
rajasree-r: Got it, updated the code. Thank you Sowmya!

PR #114, opened 2020-05-08
Solved first question

tejassjain:


deleteandearn.java
```diff
@@ -0,0 +1,31 @@
+//Time Complexity: O(NlogN)
```
sowmyadvn, line 1: We are not sorting or performing any other operations of complexity NlogN. It would be O(N) as we are iterating through the array

sowmyadvn:
Please review in-line comment Tejas. Try solving the second problem

sowmyadvn:
Please review in-line comment Tejas. Try solving the second problem

PR #139, opened 2020-06-08
Done both problems, please check comments in MinimumFalling

Shashank001122:


MinimumFalling.py
```diff
@@ -0,0 +1,13 @@
+class Solution:
+    def minFallingPathSum(self, A: List[List[int]]) -> int:
+        for i in range(1,len(A)):
+            for j in range(len(A[0])):
+                if(i-1>0 and j+1>0 and j+1<len(A[i])and i-1<len(A[i])):
```
Lohith-Aswathappa, line 5: if and elif conditions are wrong for the boundary conditions, check again. you should be able to get the correct answers

Lohith-Aswathappa:
Nice try, Shashank!
Review the inline comments

PR #152, opened 2020-06-19
Submitted for review. Completed python codes.

Abhishek1004:


AbhiniveshP:
Nice work Abhishek !!!
In delete and earn question, please explicit about what n is in O(n) as it can be either max(nums) or len(nums)

AbhiniveshP:
Nice work Abhishek !!!
In delete and earn question, please explicit about what n is in O(n) as it can be either max(nums) or len(nums)

PR #154, opened 2020-07-17
solutions to 2 problems..please review

khgupta:


samarthjj:
Good job Khushboo, please also complete the recursive solution as well!

samarthjj:
Good job Khushboo, please also complete the recursive solution as well!

Problem1.py
```diff
@@ -0,0 +1,19 @@
+#Time complexity: O(n) where n is the length of input array
```
samarthjj, line 1: The time complexity will be O(max(nums)) since the freq array is of that length!

PR #164, opened 2020-07-20
DP-3 Please review-Completed

firoz1905:


Minimum Falling Path
```diff
@@ -0,0 +1,22 @@
+Approach 1 : Brute Force Approach
+Time Complexity :O(n*2)
```
Sumesh-H, line 2: Always specify what n stands for. It will answer your question. O(n^2 + n) = O(n^2), where n is the row/col length. Space complexity is correct.

Sumesh-H (commented):
Good work, Firoz!

Sumesh-H:
Good work, Firoz!

PR #199, opened 2020-08-31

rajesh-durbha:


DeleteAndEarn.java
```diff
@@ -0,0 +1,45 @@
+// Time Complexity : O(N)
```
Sumesh-H, line 1: Where N is?

Sumesh-H (commented):
Good Job, Rajesh. Review inline comment.

Sumesh-H:
Good Job, Rajesh. Review inline comment.

DeleteAndEarn.java
rajesh-durbha: Highest point in input array

PR #234, opened 2020-09-26
Completed 2

koshyanand:


problem2.py
```diff
@@ -0,0 +1,30 @@
+# Time Complexity: O(log(N))
```
Lohith-Aswathappa, line 1: Could you explain how it's logN?

Lohith-Aswathappa (commented):
Nice try, Anand!
Check the inline comment and try the other problem as well.

Lohith-Aswathappa:
Nice try, Anand!
Check the inline comment and try the other problem as well.

PR #250, opened 2020-10-25
Delete and earn done for now- DP3

khannakanika:
Delete and earn for now. Will do another question shortly.


khannakanika:
Both questions done

PR #311, opened 2021-01-06
adding solution to dp-3

yahska-ednihs:


delete_and_earn.py
```diff
@@ -0,0 +1,32 @@
+# Time Complexity : O(n)
+# Space Complexity : O(1)
```
samarthjj, line 2: Space: O(n)

samarthjj:
Good work Akshay, Please review the inline comments.
Please also submit the recursive solution for all DP problems too.

samarthjj:
Good work Akshay, Please review the inline comments.
Please also submit the recursive solution for all DP problems too.

PR #334, opened 2021-01-31
DP3 PROBLEMS

jammulanikhil1:


kannan-mrugapphan (commented):
Good Job. Try to code the solutions recursively as well. Also please include the space complexity as well.

kannan-mrugapphan:
Good Job. Try to code the solutions recursively as well. Also please include the space complexity as well.

DP3problems/src/minimumfallingpath.java
```diff
@@ -0,0 +1,36 @@
+class minimumfallingpath {
```
aimiebell, line 1: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Class names should start with an upper case letter (NM_CLASS_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-class-names-should-start-with-an-upper-case-letter-nm-class-naming-convention).
Class names should be nouns, in mixed case with the first letter of each internal word capitalized. Try to keep your class names simple and descriptive. Use whole words-avoid acronyms and abbreviations (unless the abbreviation is much more widely used than the long form, such as URL or HTML).

PR #338, opened 2021-02-01
implemented Falling Sum and Delete & Earn

itselavia:


keshav-Gupta-IS (commented):
Good work, Akshay!! Keep it up. Yes delete and earn problem logic is not very easy to come up with. But as you move along in the cohort you will start developing such logics easily.

keshav-Gupta-IS:
Good work, Akshay!! Keep it up. Yes delete and earn problem logic is not very easy to come up with. But as you move along in the cohort you will start developing such logics easily.

PR #345, opened 2021-02-15
Solutions

DikshaWaila95:
Please provide comments on my work.

### DIFF
--- a/Problem1.py
+++ b/Problem1.py
@@ -1,0 +1,37 @@
+"""
+740. Delete and Earn
+Time Complexity - O(n)
+Space Complexity - O(10**4) worst case
+Creating a result list where index indicates the number and value at that index indicates total sum of that number in list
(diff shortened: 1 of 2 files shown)


Problem1.py
```diff
@@ -0,0 +1,37 @@
+"""
+740. Delete and Earn
+Time Complexity - O(n)
+Space Complexity - O(10**4) worst case
```
yahska-ednihs, line 4: You don't have to consider the maximum possible input. Specify your time and space complexity in term of input. if there are n inputs then space complexity would be n or n**2 depending on the algorithm. Here it is n.

Problem2.py
```diff
@@ -0,0 +1,31 @@
+"""
+931. Minimum Falling Path Sum
+Time Complexity - O(n*m)
+Space Complexity - O(1)
+Here we are finding optimal solution starting from row=1 and most optimal solution from last row will be our solution"""
+
+class Solution:
```
yahska-ednihs, line 7: Good Job! Please review inline comments.

PR #438, opened 2021-06-13
Done falling path and delete and earn

hharshitarora:
I used and top down approach rather than the bottom up approach as taught in class.


Rooppesh:
Good job! Please add the TC and SC as well!

Rooppesh:
Good job! Please add the TC and SC as well!

PR #467, opened 2021-07-05
completed optimum code

venkatagudimela:
complted optimal code please verify


Rooppesh:
Good Job with the code. Please add TC and SC along with the comments.

Rooppesh:
Good Job with the code. Please add TC and SC along with the comments.

PR #509, opened 2021-08-28
DP-3 complete please review

rajchoksi1997:


Pinak08 (commented):
Good job!
The SC for 1st problem will be O(Max element of nums). You are thinking 10001 is a number so it should be constant but as constraints of question were 10000 so you took 10001 or else you would have taken this as much so space complexity is not constant.

Pinak08:
Good job!
The SC for 1st problem will be O(Max element of nums). You are thinking 10001 is a number so it should be constant but as constraints of question were 10000 so you took 10001 or else you would have taken this as much so space complexity is not constant.

PR #525, opened 2021-09-07
Dp3 please review tc and sc of delete and earn

veeceey:


maithrreye:
Good Job on delete and earn !

maithrreye:
Good Job on delete and earn !

PR #544, opened 2021-10-07
Completed DP-3. Please review

ManjariK0308:


brunda4 (commented):
Please update the correct SC for minFalling Path and Delete and earn problem does not require sorting , it is just adding on the TC.

brunda4:
Please update the correct SC for minFalling Path and Delete and earn problem does not require sorting , it is just adding on the TC.

PR #782, opened 2022-06-21
Completed DP-3 : Please review

tiwariarjun91:


Delete_And_Run.py
```diff
@@ -1,31 +1,17 @@
 # // Time Complexity : O(n)
 # // Space Complexity : O(n)
-# // Did this code successfully run on Leetcode : No, the logic was correct but it gave time limit exceed error
```
ishika2741: class Solution:
    def deleteAndEarn(self, nums: List[int]) -> int:
        # base case
        if len(nums) == 0 or not nums:
            return -1

        m = max(nums) + 1 # built-in function to get max of the list, add +1 as list is 0 based

        arr = [0] * m    # create empty list with m elements

        for n in nums:  # O(n)
            arr[n] += n

        # once arr is constructed, we have reduced the problem to house robber problem
        dp0 = arr[0]
        dp1 = max(dp0,0+arr[1])  # at 0th index, just select nums[0], at 1st index choose max of 0th and 1st index

        for i in range(2,len(arr)): # O(max(n))
            dp1, dp0 = max(arr[i]+dp0, dp1), dp1

        return dp1  
        
        
        Try this code . It works.

ishika2741 (commented):
Good job!

ishika2741:
Good Job!

PR #798, opened 2022-07-06
complete dp3

tusharpawar611:


niharikagupta2021 (commented):
Good Job! Please add sc and tc as well

niharikagupta2021:
Good Job! Please add sc and tc as well

PR #833, opened 2022-08-11
Completed both the ques

as-DSAPrep:


LC740_DeleteandEarn.py
```diff
@@ -0,0 +1,22 @@
+
+"""
+TC = O(n)
+SC = O(1)
```
daredevilteja, line 4: SC is O(max(nums))

daredevilteja:
Good Work! Keep it up

PR #888, opened 2022-10-10
Completed DP3

gourishpisal08:


delete_earn.py
```diff
@@ -0,0 +1,15 @@
+#TC O(n)
+#SC O(1)
```
daredevilteja, line 2: SC is O(n)

daredevilteja:
Good Work! Keep it up

PR #897, opened 2022-10-18
DP-3: Please review

pavansai1:


falling-path-sum.java
```diff
@@ -0,0 +1,54 @@
+// TC: O(n*r)
+// SC: O(r)
```
daredevilteja, line 2: SC is O(n*n)

delete-earn.java
```diff
@@ -0,0 +1,56 @@
...
+Then for the elements present in the given array put the ele  freq[i] in the new array and for the
+ rest put 0. Then now you can safely assume that it is house robber problem.
+
+*/
+
+// Approach: Here we are sorting the array and will start from the first element. the two while loops
+// below will add all the current element to the sum and curr+1 ele to the sum and then we move to
+// the next index. Observe the "nextidx" parameter.
+
+// TC: O(n)
```
daredevilteja, line 15: TC is O(nlogn) because of sorting

daredevilteja:
Good Work! Keep it up

PR #900, opened 2022-10-21
DP-3 changes completed. Please review

jadhpradeep:


DeleteAndEarnMaxPoints.java
```diff
@@ -0,0 +1,47 @@
+//TC: O(n)
+//SC: O(1)
```
daredevilteja, line 2: SC is O(k) because of freq array

daredevilteja:
Good Work! Keep it up

PR #1071, opened 2023-09-13
Completed: Min Falling Path

rinkyfulwani:


pvk3081 (commented):
Second problem is missing

pvk3081:
Second problem is missing

rinkyfulwani:
> Second problem is missing

Completed second problem Delete and earn

PR #1419, opened 2025-05-28
delete and earn solution

kedarbhusari:


dalvirushikesh:
Great work! Please submit solution for second question.

PR #1456, opened 2025-07-11
DP-3 Min falling and Delete Earn

botlamanoj29:
Both the Leet code problem submitted.

### DIFF
--- a/DeleteEarn.py
+++ b/DeleteEarn.py
@@ -1,0 +1,37 @@
+# Time Complexity : It is O(N) since looping through the list.
+# Space Complexity : It is O(N) since we are creating a new list.
+# Did this code successfully run on Leetcode :
+# Any problem you faced while coding this :
+
+from typing import List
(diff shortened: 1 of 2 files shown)


super30admin:
- **Strengths**: 
  - The student attempted to use dynamic programming, which is the right approach for this problem.
  - The helper function `findTheMinNeighbour` is a good idea but needs to be integrated correctly into the main logic.
  
- **Areas for Improvement**:
  - The logic for calculating the minimum falling path sum is incorrect. The student should revisit the problem and understand how to correctly accumulate the minimum path sum from the bottom up.
  - The edge case handling (e.g., single-element matrix) needs to be corrected.
  - Variable names should be more descriptive to improve readability.
  - The space complexity claim is incorrect; modifying the input matrix in place does not make it O(1) space if the input is considered mutable.
  - The solution should include more comments to explain the approach and logic.

PR #1464, opened 2025-08-03
 Done Problem1

RahulGundapaneni:


super30admin:
- The student's solution is for a different problem (Delete and Earn) and does not address the problem of finding the minimum falling path sum in a matrix.
- The approach and logic used in the solution are correct for the Delete and Earn problem, but they are not applicable to the minFallingPathSum problem.
- The student should ensure they are solving the correct problem before submitting their solution.
- The code quality and structure are good, and the comments are helpful, but they should be aligned with the correct problem.

RahulGundapaneni:
Hi Sir, I solved Problem 1 from the given Repository. Could you please correct me if I am wrong?

In ReadMe, it has 2 problems: Problem1 and Problem2
https://github.com/super30admin/DP-3

PR #1468, opened 2025-08-05
completed DP-3

Lavanyarekkala:
Added the code for minfallingpath sum too


super30admin:
Strengths:
- The use of a single dp array and tracking previous values optimizes space complexity.

Areas for Improvement:
- The handling of the `prev` variable could be simplified to avoid the need for resetting it when `j == len`.
- The edge case for a single-row matrix could be handled more elegantly within the main loop rather than as a separate check at the end.
- Consider initializing the `dp` array directly with the last row of the matrix to avoid the separate loop for initialization.

PR #1469, opened 2025-08-06
Done DP-3

mlokesh22:


super30admin:
Strengths:

Areas for Improvement:
- Consider optimizing the space complexity by using a single row for DP storage instead of a full matrix. This would reduce the space complexity from O(n^2) to O(n).
- The edge case handling for the first row could be simplified by initializing the DP table directly with the first row values and then iterating from the second row.

PR #1482, opened 2025-09-01
Implemented DP-3

johar-inderpreet:
Implemented DP-3


super30admin:
- Strengths:
  - The student has correctly identified the dynamic programming approach and implemented it efficiently.
  - The solution is well-commented and easy to understand.
  - The in-place modification of the matrix is a clever optimization that saves space.
- Areas for Improvement:
  - While the solution is correct, it might be worth mentioning in comments why modifying the input matrix is acceptable (if the problem constraints allow it).
  - The variable names could be more descriptive (e.g., `n` could be `size` or `matrixSize`).

PR #1490, opened 2025-09-08
Create earnanddelete

Harsha-Ramesh-Naik:


super30admin:
- The student's solution is for a different problem (deleteAndEarn) and does not address the given problem (minimum falling path sum in a matrix).
- The student should carefully read and understand the problem statement before attempting to solve it.
- The code quality is good, but it's important to ensure the solution matches the problem requirements.
- For the correct problem, the student should consider using dynamic programming to store intermediate results and avoid recalculating them, similar to the reference solution.

PR #1504, opened 2025-11-26
Completed DP-3

pranavs0133:


super30admin:
Strengths:

Areas for Improvement:
- The space complexity can be optimized further by using a 1D array for the DP table instead of a 2D array, reducing it from O(m*n) to O(n).
- The edge case handling could be more explicit, such as checking for null input or empty matrix.

PR #1519, opened 2026-01-10
Delete and earn. Minimum falling path sum

paridhimalviya:
Problem1: (https://leetcode.com/problems/delete-and-earn/)

Problem2 (https://leetcode.com/problems/minimum-falling-path-sum/)


super30admin:
Strengths:
- The student demonstrates a good understanding of dynamic programming concepts, both in the brute-force and tabulation approaches.
- The code is well-commented and structured, making it easy to follow.
- The student correctly identifies the need for dynamic programming to optimize the solution.

Areas for Improvement:
- The brute-force solution should be generalized to handle matrices of any size, not just 3x3.
- The tabulation approach has a bug in the final step where it doesn't check all columns for the minimum value. This should be fixed by iterating over all columns in the first row.
- Consider using a rolling array in the tabulation approach to optimize space complexity further.
- The helper function in the brute-force approach could be simplified by handling all columns uniformly instead of using conditional checks for each column.

PR #1521, opened 2026-01-16
Added Solution to both the problems

joshiatharv1:


super30admin:
- The student has done an excellent job in solving the problem efficiently and correctly. The use of dynamic programming to reduce both time and space complexity is commendable.
- The comments and variable names make the code easy to understand.
- One minor suggestion would be to add a brief explanation of the approach at the beginning of the method, which could help in quickly understanding the logic without diving into the code details.
- The student should ensure that the file names and problem descriptions match the actual problem being solved, as the first file (Problem1.java) seems to be for a different problem (Delete and Earn).

PR #1541, opened 2026-05-23
DP3

YogeshPardeshi:


super30admin:
### Least Falling Path Sum (Problem1.java)
The major issue is that you've solved a completely different problem. The "Least Falling Path Sum" problem requires finding minimum sum paths through a matrix moving only down, down-left, or down-right. Your solution about deleting numbers to earn points is for a different problem entirely.

If you want to solve the correct problem:
- Use dynamic programming starting from the last row
- For each cell, calculate the minimum sum to reach that cell from any valid starting position
- The answer is the minimum value in the first row after computing the DP

Your code quality is good for the problem you attempted, but you must ensure you're solving the right problem before optimizing.

VERDICT: NEEDS_IMPROVEMENT

---

### Delete and Earn (Problem2.java)
- The student appears to have submitted code for a completely different problem
- The implementation of "Minimum Falling Path Sum" is actually well-structured with proper handling of edge cases (first and last columns)
- However, this solution is completely irrelevant to the "Delete and Earn" problem
- The student needs to understand and implement the correct problem: Delete and Earn requires:
  1. Counting the sum of all occurrences of each number
  2. Applying the house robber DP pattern where dp[i] = max(dp[i-1], arr[i] + dp[i-2])
- The code quality is good (proper indentation, clear logic flow) but for the wrong problem

VERDICT: NEEDS_IMPROVEMENT

